PR #21, opened 2025-01-06
Update README.md

Jeadie:
pg_source is not in [`spiceai/fed-demo`](https://spice.ai/spiceai/fed-demo) spicepod

### DIFF
--- a/federation/README.md
+++ b/federation/README.md
@@ -59,9 +59,6 @@ select * from s3_source;
 -- Query the accelerated S3 source
 select * from s3_source_accelerated;
 
--- Query the federated PostgreSQL source
-select * from pg_source;
-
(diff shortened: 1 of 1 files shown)


federation/README.md
```diff
@@ -70,8 +67,6 @@ select * from dremio_source_accelerated;
 
 -- Perform an aggregation query that combines data from S3, PostgreSQL, and Dremio
```
Sevenannn: The comments need to be updated to remove PostgreSQL

Sevenannn:
```suggestion
-- Perform an aggregation query that combines data from S3 and Dremio
```

Sevenannn (approved):

PR #138, opened 2025-04-15
Fix the spice-qa workflow

Sevenannn:
## 🗣 Description

Use the spice cli and spice chat command to get response from SCP timmy in spice qa workflow
The fixed workflow can run file connector recipe e2e and upload the response as artifact: https://github.com/spiceai/cookbook/actions/runs/14486147753

## 🤔 Concerns

The matrix strategy will face the challenge of different recipes running in parallel within the same timmy sandbox, which will conflict with each other. Future recipes to be added should be running in a sequential manner, unless SCP supports different timmy sandboxes for different recipes.


.github/workflows/spice-qa.yml
```diff
@@ -78,8 +65,17 @@ jobs:
             exit 1
           fi
 
+      - name: Retrieve response from streaming result
+        shell: bash
+        env:
+          SPICE_API_KEY: ${{ secrets.SPICE_API_KEY }}
+        run: |
+          input_file="api_responses/${{ env.RECIPE_FILE_PATH }}_stream.log"
```
lukekim: If we need to write external code to do this, we should write it in the spice CLI, not make another program.

Sevenannn:
- Closed in favor of: https://github.com/spiceai/cookbook/pull/146

PR #179, opened 2025-06-09
Improve Azure OpenAI Models recipe

sgrebnov:
## 🗣 Description

1. Update search output format to match the latest
1. Include `INFO runtime: All components are loaded. Spice runtime is ready!` trace as important part to understand when runtime is ready


Jeadie (commented):
These search changes are in  https://github.com/spiceai/cookbook/pull/173

PR #196, opened 2025-06-29
feat: SpiceQA display success / failure status according to the test report

Sevenannn:
## 🗣 Description

The test report of SpiceQA now contains a consistent line containg the success / failure status, in the following form:
```
✅ QA passed (90% confidence)
or 
❌ QA failed (60% confidence)
```
Determine the success / failure status of a recipe according to the test report line in the workflow, so that the workflow is red(failed) when a recipe failed, and the workflow is green (passed) when a recipe succeeded, eliminating manual check on the recipe files.

E2E run succeeded in CI:
https://github.com/spiceai/cookbook/actions/runs/16013588017/job/45175994191

A sample run using synthetic log data:
https://github.com/spiceai/cookbook/actions/runs/15958376658


.github/workflows/spice-qa.yml
```diff
@@ -68,3 +68,22 @@ jobs:
...
+            grep -oE '✅ QA passed \([0-9]+% confidence\)|❌ QA failed \([0-9]+% confidence\)' \
+              "api_responses/${{ env.RECIPE_FILE_PATH }}.log" \
+            | sed -E \
+                -e 's/✅ QA passed \(([0-9]+)% confidence\)/passed \1/' \
+                -e 's/❌ QA failed \(([0-9]+)% confidence\)/failed \1/'
+          )
+          echo "QA Status: $qa_stat"
+          echo "QA Score: $qa_score"
+
+          if [[ "$qa_stat" == "failed" ]]; then
```
kczimm: Is there a chance that `qa_stat` is a value besides "passed" or "failed"? maybe empty? The current script defaults to success unless "failed" is found. Perhaps it would be better to default to return a non-zero value unless an explicit "success" is found.

Sevenannn: Updated, thanks!

sgrebnov (approved):

kczimm (approved):

PR #200, opened 2025-07-03
Update README.md

edawg2:


edawg2:
I was also not sure if we want to update to Llama 4 now that it is released.

Advayp (commented):
`spicepod.yml` is fine here, since it's actually in the `llama/` directory. I don't think we should change it to `.yaml` because technically it's not in the directory.

PR #241, opened 2025-07-31
Validate redshift

mach-kernel:
## 🗣 Description

Adds a Redshift round-trip cookbook. To make setting up test data easier, we use our S3 TPCH dataset and Redshift as an acceleration target to do writes, then read them back out with the PG dataset connector.


redshift/README.md
```diff
@@ -0,0 +1,225 @@
+# redshift-tpch-roundtrip
+
+Round-trip TPCH through Redshift with Spice
```
lukekim: Set up the tutorial with more context for the user about what it's about, what it does, etc.

redshift/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+#### Create Redshift Cluster
```
lukekim: Fix MD headings. (Add Prettier or equivalent)

redshift/README.md
```diff
@@ -0,0 +1,225 @@
...
+  --parameters ParameterKey=MasterUsername,ParameterValue=admin \
+               ParameterKey=MasterUserPassword,ParameterValue=hGG3ellothere$$$$ \
+               ParameterKey=DatabaseName,ParameterValue=dev
+```
+
+#### Write data into Redshift as accelerator target
+
+Start Spice:
+
+```bash
```
sgrebnov, line 27: Nit: avoid using extra characters so it can be copy-pasted into terminal for testing and easier for SpiceQA (AI), `spice run` is a more standardized way for user to use spice

```bash
cd write
spice run
```

sgrebnov: Please also consider adding spice output/trace examples where it is  applicable/makes sense.

redshift/README.md
```diff
@@ -0,0 +1,225 @@
...
+count_order    | 10
+```
+
+#### Read data from Redshift via PostgreSQL connector
+
+Start Spice:
+
+```bash
+$ pushd read
+$ spiced
```
Advayp: Should this be `spice run`?

redshift/README.md
```diff
@@ -0,0 +1,225 @@
...
+```
+
+Validate that the correct tables have been made, look at the schema of one or two:
+```bash
+$ spiced --repl
```
Advayp: Similarly, should this be `spice sql`?

redshift/README.md
```diff
@@ -0,0 +1,225 @@
...
+| l_shipdate      | Date32            | YES         |
+| l_commitdate    | Date32            | YES         |
+| l_receiptdate   | Date32            | YES         |
+| l_shipinstruct  | Utf8              | YES         |
+| l_shipmode      | Utf8              | YES         |
+| l_comment       | Utf8              | YES         |
++-----------------+-------------------+-------------+
+```
+
+Run a TPCH query:
```
Advayp: Consider separating SQL + output, so that the SQL is easier to copy (and also has syntax highlighting)

redshift/write/spicepod.yaml
Advayp, line 1: Very nitpicky, but you could use [YAML anchors](https://medium.com/@kinghuang/docker-compose-anchors-aliases-extensions-a1e4105d70bd) here to avoid repetition (applies to the other Spicepod too)

redshift/README.md
mach-kernel: Done x2

lukekim (approved):

PR #495, opened 2026-06-28
Fix snowflake/dml recipe: phantom tv_shows_ro output and broken install link

claudespice:
## What the recipe says

`snowflake/dml/README.md` Step 3 expected output shows a third dataset registering and a 4-component ready count:

```
... Dataset tv_shows_ro registered (snowflake:SPICE_DEMO.PUBLIC."TV_SHOWS") ...
... Dataset tv_shows registered (snowflake:SPICE_DEMO.PUBLIC."TV_SHOWS") ...
... Dataset load summary (after 30s): 4/4 ready, 0 unhealthy, 0 still initializing.
```

## What the recipe actually ships

The shipped `snowflake/dml/spicepod.yaml` defines only **two** datasets — `tvmaze_shows_raw` and `tv_shows` — plus one worker. There is no `tv_shows_ro` dataset; it is a leftover from an earlier revision of the recipe that had a read-only mirror dataset.

The `4/4 ready` count is also wrong. The load summary counts **datasets only**, not workers — verified against `spiceai/spiceai` at tag `v2.0.1`, `crates/runtime/src/init/dataset.rs:231-255` (`get_dataset_statuses()` → `{ready}/{total}`). With two datasets the line is `2/2 ready`.

## What changed

- Removed the `Dataset tv_shows_ro registered ...` line (that dataset is not in the spicepod).
- Corrected the load summary from `4/4 ready` to `2/2 ready` to match the two shipped datasets.

### DIFF
--- a/snowflake/dml/README.md
+++ b/snowflake/dml/README.md
@@ -6,7 +6,7 @@ Works with `v2.0+`
 
 ## Pre-requisites
 
-- Spice `v2.0+` — [Install Spice](https://docs.spiceai.org/getting-started/installation)
+- Spice `v2.0+` — [Install Spice](https://docs.spiceai.org/getting-started)
 - [A Snowflake account](https://signup.snowflake.com/)
 
 ## Step 1. Create the destination table in Snowflake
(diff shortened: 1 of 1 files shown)


claudespice:
Added a second fix to this branch: the Pre-requisites install link pointed to `https://docs.spiceai.org/getting-started/installation` (returns **404**); corrected to `https://docs.spiceai.org/getting-started` (`200`, same canonical target as #487). Same recipe, so folding it into this PR rather than opening a separate one.

Jeadie (approved):

Jeadie (approved):

PR #530, opened 2026-07-07
Add postgres/mysql CDC cookbooks

krinart:
## 🗣 Description

Add postgres/mysql CDC cookbooks

### DIFF
--- a/mysql/cdc/README.md
+++ b/mysql/cdc/README.md
@@ -1,6 +1,6 @@
 # MySQL CDC (Binlog Replication)
 
-Works with `v2.0+`
+Works with `v2.2.0+`
 
 This recipe demonstrates how to stream real-time changes from a MySQL table into Spice using native Change Data Capture (CDC) over the MySQL [binary log](https://dev.mysql.com/doc/refman/8.0/en/binary-log.html). Inserts, updates, and deletes propagate automatically to the Spice accelerator — no Debezium or Kafka required.
 
--- a/mysql/cdc/README.md
+++ b/mysql/cdc/README.md
@@ -1,0 +1,210 @@
+# MySQL CDC (Binlog Replication)
+
+Works with `v2.0+`
+
+This recipe demonstrates how to stream real-time changes from a MySQL table into Spice using native Change Data Capture (CDC) over the MySQL [binary log](https://dev.mysql.com/doc/refman/8.0/en/binary-log.html). Inserts, updates, and deletes propagate automatically to the Spice accelerator — no Debezium or Kafka required.
+
+Spice reads the binlog directly and applies row-level changes by primary key. The current binlog file and position are checkpointed in a client-side sidecar table so streaming resumes from where it left off after a restart.
+
+## Prerequisites
+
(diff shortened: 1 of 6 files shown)


mysql/cdc/README.md
```diff
@@ -0,0 +1,210 @@
+# MySQL CDC (Binlog Replication)
+
+Works with `v2.0+`
```
bjchambers: Won't binlog replication require the *next* version since it isn't implemneted yet?

mysql/cdc/README.md
```diff
@@ -0,0 +1,210 @@
...
+docker exec mysql-cdc mysql -uroot -pspice \
+  -e "SHOW VARIABLES WHERE Variable_name IN ('log_bin','binlog_format','binlog_row_image');"
+```
+
+```console
++-------------------+-------+
+| Variable_name     | Value |
++-------------------+-------+
+| binlog_format     | ROW   |
+| binlog_row_image  | FULL  |
```
bjchambers, line 44: Note for future improvement -- this sends us a full image of the before and after row for each change. At some point, it may be better to support `MINIMAL`, but that would require us using the primary key to lookup the previous value to merge with whatever changes we receive. Definitely not something to change today, just flagging it.

postgres/cdc/README.md
```diff
@@ -0,0 +1,203 @@
+# PostgreSQL CDC (Logical Replication)
+
+Works with `v2.0+`
```
bjchambers, line 3: Same question re: version

postgres/cdc/README.md
```diff
@@ -0,0 +1,203 @@
...
+
+This recipe demonstrates how to stream real-time changes from a PostgreSQL table into Spice using native Change Data Capture (CDC) over [logical replication](https://www.postgresql.org/docs/current/logical-replication.html) (WAL streaming). Inserts, updates, and deletes propagate automatically to the Spice accelerator — no Debezium or Kafka required.
+
+Spice reads the write-ahead log (WAL) through a replication slot and applies row-level changes by primary key. The dataset first bootstraps from a consistent snapshot, then streams live changes from the slot so replication resumes from where it left off after a restart.
```
bjchambers: I believe postgres runs a walsender that *reads the WAL and sends Spice* the changes. Not sure about the mysql case. But, I think it's basically equivalent (although, each slot does increase load on postgres)?

bjchambers (approved):

postgres/cdc/README.md
krinart: Thanks, updated

mysql/cdc/README.md
krinart: Updated

mysql/cdc/README.md
krinart: Noted - makes sense

postgres/cdc/README.md
krinart: Postgres WAL CDC is already implemented - but I won't this PR until the next release to make sure we have mysql support

PR #560, opened 2026-08-01
Fix pre-v2 CLI transcripts in `spice search` and `spice trace` output

claudespice:
## Summary

The Spice CLI was rewritten in Rust (spiceai/spiceai#9061, Jan 2026). Since then — and in every currently installable release, including stable `v2.1.2` — `spice search` and `spice trace` render through `TableOutput`, which uses comfy-table's `NOTHING` preset: **no borders**, different column headers, `OK`/`ERR` statuses, and `Time: <seconds> seconds.` timing.

Three recipes still show the pre-rewrite Go CLI output, so a reader following them today sees something that looks nothing like their terminal.

| Recipe | What the README showed | What `v2.1.2` prints |
| --- | --- | --- |
| `docker` | `Rank 1, Score: 89.7, Datasets [spice.public.films]` line output; `Time: 32ms.` | borderless `Rank / Match / Score / Dataset` table; score `0.8975`; `Time: 0.032 seconds.` |
| `azure_openai` | bordered `+---+` table, headers `Key (path)` / `Dataset(s)`; `Time: 650ms.` | borderless table, headers `Key` / `Dataset`; `Time: 0.650 seconds.` |
| `text-to-sql` | `TREE STATUS DURATION SPANID` headers, check-mark status column | `Tree / Status / Duration / Span ID` headers, `OK` status |

The `docker` score is the clearest break: the README printed `89.7` while the API response immediately below it — for the same match, same query — reports `"_score": 0.8974827855112448`. The old CLI displayed the score scaled ×100; the current one prints the raw score to four decimals.

## Verified against

`spiceai/spiceai` at `v2.1.2`:

- [`bin/spice/src/output/table.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/bin/spice/src/output/table.rs) — `TableOutput` loads `presets::NOTHING` (borderless; one space of padding per cell).
- [`bin/spice/src/commands/search.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/bin/spice/src/commands/search.rs) — headers are `["Rank", "Match", "Score", "Dataset"]`, or `["Rank", "Key", "Match", "Score", "Dataset"]` when results carry a primary key; score is formatted `{:.4}`; the footer is `Time: {:.3} seconds. N results.`
- [`bin/spice/src/commands/trace.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/bin/spice/src/commands/trace.rs) — headers are `["Tree", "Status", "Duration", "Span ID"]` (+ `Input`/`Output`), status is `OK`/`ERR`, duration is `{:>8.2}ms`.

The `azure_openai` table keeps its `Key` column (its API response includes `primary_key`); the `docker` table has none (its API response has no `primary_key`).

## Evidence

Static review — the recipes need Docker, Azure OpenAI, and OpenAI credentials, so they were not run.

The layout was reproduced from the renderer and checked byte-for-byte against a transcript already in this repo that *was* captured on v2.x — [`mcp/README.md`](https://github.com/spiceai/cookbook/blob/trunk/mcp/README.md) shows real `spice trace ai_chat` output, and a re-render of those rows through the `NOTHING` preset rules matches it exactly. Every value in the corrected blocks is carried over from the transcript it replaces; only the layout, headers, status column, score scale, and timing format changed.

Two judgement calls worth a reviewer's eye:

1. **`docker` is truncated to three results (`...`).** The old transcript listed ten, but only the top three have exact scores documented — they appear in the recipe's own API response block. Rendering ranks 4–10 would have meant inventing four-decimal precision from one-decimal percentages, so the transcript stops at three and marks the truncation.
2. **`text-to-sql`'s trace is now flat, with no `├──` glyphs.** That is genuinely what the CLI prints today: in `recurse_through_tree`, `new_indent` returns `String::new()` whenever `indent` is empty, so the indent never grows past the root and the connector branch is unreachable. The `mcp` recipe's captured v2.x output is flat for the same reason. This looks like a CLI regression rather than intended behaviour — happy to revert this hunk if the tree rendering is going to be restored upstream instead.


grokspice:
@copilot review

grokspice:
@copilot review

lukekim (approved):

PR #561, opened 2026-08-01
Fix json_strings recipe: refresh output captured on Spice v1.x

claudespice:
## Summary

Every console block in this recipe was captured on a v1.x runtime in December 2024. I re-ran the recipe end to end against stable **v2.1.2** and replaced the blocks with what it actually prints today. Values and row counts are unchanged — only the output shape drifted.

- **SQL results carry a data-type row.** Since v2.0.0 the SQL REPL formats results with `format_batches_with_types`, which prints the column types under the names and centers the header. The recipe showed the older untyped, left-aligned header.
- **`describe` returns four columns.** It gained a leading `table_schema`; the recipe showed three.
- **The `->` derived column name changed.** `properties->'inventory'` now projects as `properties -> 'inventory'`, not `products.properties -> Utf8("inventory")`.
- **The startup log is stale.** It advertised a metrics listener on `127.0.0.1:9090` and an OpenTelemetry listener on `127.0.0.1:50052` — neither is emitted by a default v2.1.2 run — and `Initialized results cache;` is now three separate `sql`/`search`/`embeddings` cache lines.

## Verified against

`spiceai/spiceai` at `v2.1.2` — [`crates/repl/src/pretty.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/crates/repl/src/pretty.rs) (`format_batches_with_types`, unconditional in non-expanded mode) and [`crates/repl/src/lib.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/crates/repl/src/lib.rs#L994).

## Evidence

Ran it. Official `v2.1.2` release binaries, this recipe's own `spicepod.yaml`, this recipe's own queries in README order:

```console
$ spice sql
sql> describe products;
+--------------+-------------+-----------+-------------+
| table_schema | column_name | data_type | is_nullable |
|    varchar   |   varchar   |  varchar  |   varchar   |
+--------------+-------------+-----------+-------------+
| public       | id          | Int64     | YES         |
| public       | name        | Utf8      | YES         |
| public       | properties  | Utf8      | YES         |
+--------------+-------------+-----------+-------------+

Time: 0.00167675 seconds. 3 rows.
```

Every table in the diff is pasted from that session (the per-query `Time:` lines are dropped, matching the recipe's existing style).

## Note for maintainers

The missing type row is not specific to this recipe — **74 of the cookbook's READMEs** show the pre-v2.0 untyped header. Two recipes captured on v2.x (`catalogs/unity_catalog`, `elasticsearch/connector`) already show the typed form, so the repo is currently inconsistent. This PR fixes one recipe that runs locally with no Docker or secrets; the rest need a run each to regenerate honestly. Happy to work through them if that is wanted — or to hold off, if the type row is considered noise that should change on the CLI side instead.

### DIFF
--- a/json_strings/README.md
+++ b/json_strings/README.md
@@ -66,44 +66,55 @@ spice run
 Result:
 
 ```console
-2024/12/22 16:15:42 INFO Checking for latest Spice runtime release...
-2024/12/22 16:15:42 INFO Spice.ai runtime starting...
-2024-12-23T00:15:42.923112Z  INFO runtime::init::dataset: Initializing dataset products
-2024-12-23T00:15:42.925812Z  INFO runtime::metrics_server: Spice Runtime Metrics listening on 127.0.0.1:9090
-2024-12-23T00:15:42.926466Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
-2024-12-23T00:15:42.929730Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:8090
-2024-12-23T00:15:42.936005Z  INFO runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052
-2024-12-23T00:15:42.937855Z  INFO runtime::init::dataset: Dataset products registered (file://tshirts.csv).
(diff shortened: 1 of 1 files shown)


grokspice:
@copilot review

grokspice:
@copilot review

lukekim (approved):

PR #564, opened 2026-08-02
Fix localpod recipe: child dataset never picked up parent data

claudespice:
## Summary

The `localpod` recipe does not produce its documented result on v2.1.2. After running `./generate_data.sh`, the README says both datasets return 1,000 rows. In reality the parent reaches 1,000 and **the localpod child stays at 0 indefinitely** — which is precisely the behaviour the recipe exists to demonstrate.

**Why.** The child has no refresh trigger at all:

1. Localpod refresh synchronization fails during startup. The runtime logs (at DEBUG) `Could not synchronize refreshes with parent table time_series. Parent table is not an accelerated table.`, and the success line the README shows — `Localpod dataset local_time_series synchronizing refreshes with parent table time_series` — never appears.
2. The shipped `spicepod.yaml` gives the child no `refresh_check_interval`, so with synchronization skipped it refreshes once at boot and never again. Over a 2-minute run the parent refreshed 12 times and the child exactly twice (both at t=0).

**The fix.** Add `refresh_check_interval: 10s` to the child. This is not an invention — it is what the README's own expected output has always implied was configured:

```
INFO runtime: Dataset local_time_series registered (localpod:time_series), acceleration (duckdb:file, 10s refresh), ...
```

The shipped spicepod produces `acceleration (duckdb:file)` with no interval; with the setting restored the runtime prints `acceleration (duckdb:file, 10s refresh)`, matching the README exactly, and the recipe produces its documented 1,000 / 1,000 result.

Also replaced the run transcripts, captured in **October 2024** on a v1.x runtime, with real v2.1.2 output. They contained four lines that no longer exist: `runtime::metrics_server: Spice Runtime Metrics listening`, `runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052`, `Initialized results cache;` (now three separate `runtime::init::caching` lines), and eight `Tool [...] ready to use` lines. Dataset registration also moved from `runtime:` to `runtime::init::dataset`, and SQL results now render with a data-type row.

## Verified against

spiceai/spiceai at `v2.1.2` — [`crates/runtime/src/datafusion/mod.rs`](https://github.com/spiceai/spiceai/blob/v2.1.2/crates/runtime/src/datafusion/mod.rs#L3355-L3374): the `synchronize_with` helper emits the "Parent table is not an accelerated table" debug line and returns early, skipping the `tracing::info!` sync confirmation.

## Evidence

Ran the recipe as written on v2.1.2. Before the fix — parent loads, child does not:

```console
$ ./generate_data.sh && sleep 35
$ printf 'SELECT COUNT(*) FROM time_series;\nSELECT COUNT(*) FROM local_time_series;\n' | spice sql
| 1000     |     <- time_series
| 0        |     <- local_time_series
```

Refresh counts over the same run — the child never refreshes after boot:

```console
$ grep refresh_task spice.log | grep -c "dataset time_series"        # 12
$ grep refresh_task spice.log | grep -c "dataset local_time_series"  # 2  (both at t=0)
```

The synchronization failure is deterministic, 3/3 clean starts, and is **not** specific to the Arrow accelerator — it reproduces with `engine: duckdb` on the parent too:

```console
run 1: Could not synchronize refreshes with parent table time_series. Parent table is not an accelerated table.
run 2: Could not synchronize refreshes with parent table time_series. Parent table is not an accelerated table.
run 3: Could not synchronize refreshes with parent table time_series. Parent table is not an accelerated table.
```

After the fix — the child registers with the interval the README documents, and propagation works:

```console
INFO runtime::init::dataset: Dataset local_time_series registered (localpod:time_series), acceleration (duckdb:file, 10s refresh), results cache enabled. duration_ms=3
...
INFO runtime::accelerated_table::refresh_task: Loaded 1,000 rows (24.00 B) for dataset time_series in 4ms.
INFO runtime::accelerated_table::refresh_task: Loaded 1,000 rows (24.00 B) for dataset local_time_series in 15ms.

$ printf 'SELECT COUNT(*) FROM time_series;\nSELECT COUNT(*) FROM local_time_series;\n' | spice sql
| 1000     |
| 1000     |
```

Every transcript in the updated README is copied from this run.

## Note for reviewers

This PR restores the recipe's documented outcome, but the underlying synchronization failure looks like a **runtime bug worth filing separately**. The parent is plainly an accelerated table — the runtime logs `acceleration (arrow, 15s refresh)` for it moments earlier — yet the downcast to `AcceleratedTable` in `synchronize_with` fails. The code comment there specifically anticipates the bare-`AcceleratedTable` (Arrow) case, so something else is now wrapping the parent's provider. Since the failure is logged at DEBUG and the child still registers successfully, a user sees no warning at all — the recipe just quietly returns stale data.

If synchronization is repaired upstream, the added `refresh_check_interval` remains harmless (the child simply also polls on its own schedule), but the recipe would then be a better demo of synchronized refreshes without it.

### DIFF
--- a/localpod/README.md
+++ b/localpod/README.md
@@ -25,6 +25,7 @@ datasets:
       enabled: true
       engine: duckdb
       mode: file
+      refresh_check_interval: 10s
 ```
 
 :::note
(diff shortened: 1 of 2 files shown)


grokspice:
@copilot review

grokspice:
@copilot review

lukekim (approved):

PR #568, opened 2026-08-05
Fix acceleration recipes: spiceai/quickstart no longer sets refresh_check_interval

claudespice:
## Summary

Both `acceleration/data-refresh` and `acceleration/cron` build on `spice add spiceai/quickstart`,
and both assume that Spicepod configures a 10-second refresh interval. It doesn't — the pod on
spicerack.org is now just:

```yaml
acceleration:
  enabled: true
```

**What the recipes said → what actually happens:**

1. `data-refresh` Step 2 instructs the reader to "Remove the line `refresh_check_interval: 10s`
   to prevent automated refreshes". There is no such line to remove, so the step can't be
   followed as written. Replaced with the two edits that *are* needed (`refresh_mode: full` and
   `refresh_sql`) plus a note that no automated refresh needs disabling.
2. Both Step 1 transcripts show `acceleration (arrow, 10s refresh)`; the runtime logs
   `acceleration (arrow)`.
3. Step 3's "as determined by `refresh_check_interval`" no longer applies — the manual trigger
   that follows is the only thing that applies the new `refresh_sql`.

While rewriting those transcripts from a live run, also corrected:

- `Updating accelerated dataset taxi_trips...` — this log line does not exist in v2.x
  (`git grep "Updating accelerated dataset" v2.1.3` returns nothing); a config change now logs
  `Loading data for dataset …` followed by `Dataset … registered`.
- The metrics (`:9090`) and OpenTelemetry (`:50052`) listener lines, which are off by default.
- `spice sql` output: aggregate column names print lowercase (`avg(...)`, not `AVG(...)`) and
  each table includes a type row.

Row counts and values in the recipe were all confirmed correct and are unchanged.

## Verified against

spiceai/spiceai at `v2.1.3` (current stable), plus the live `spiceai/quickstart` Spicepod
fetched by `spice add`.

## Evidence

Ran `data-refresh` end to end on v2.1.3. Step 1:

```console
2026-08-05T12:06:19.784484Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (arrow), results cache enabled.
2026-08-05T12:06:23.126886Z  INFO runtime::accelerated_table::refresh_task: Loaded 2,964,624 rows (399.38 MiB) for dataset taxi_trips in 3s 340ms.
```

```console
sql> select avg(passenger_count) from taxi_trips;
+---------------------------------+
| avg(taxi_trips.passenger_count) |
|             float64             |
+---------------------------------+
| 1.3392808966805005              |
+---------------------------------+
```

Step 2 (after adding `refresh_sql`) and Step 3 (PATCH + trigger) both produced the documented
row counts — 405,103 then 91,262 — and averages of 2.0 then 3.0.

`acceleration/cron` was also run end to end: with `refresh_cron: "*/30 * * * * *"` the dataset
refreshed at `:30` and `:00` exactly as documented, so only its Step 1 transcript needed the fix.

### DIFF
--- a/acceleration/cron/README.md
+++ b/acceleration/cron/README.md
@@ -27,14 +27,12 @@ The Spice Runtime will start and the `taxi_trips` dataset included in the `spice
 
 ```console
 Spice.ai runtime starting...
-2024-08-26T18:43:28.915833Z  INFO runtime::metrics_server: Spice Runtime Metrics listening on 127.0.0.1:9090
-2024-08-26T18:43:28.915869Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
-2024-08-26T18:43:28.915925Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:8090
-2024-08-26T18:43:28.921589Z  INFO runtime::opentelemetry: Spice Runtime OpenTelemetry listening on 127.0.0.1:50052
-2024-08-26T18:43:29.115877Z  INFO runtime: Initialized results cache; max size: 128.00 MiB, item ttl: 1s
-2024-08-26T18:43:29.636542Z  INFO runtime: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (arrow, 10s refresh), results cache enabled.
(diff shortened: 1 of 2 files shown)


grokspice:
@copilot review

lukekim (approved):

PR #573, opened 2026-08-07
Fix stale uppercase aggregate column headers in query output

claudespice:
## Summary

DataFusion names aggregate output columns in lowercase, so

```sql
select avg(total_amount), avg(tip_amount), count(1), passenger_count from taxi_trips group by passenger_count order by passenger_count asc;
```

produces headers `avg(taxi_trips.total_amount)` / `count(Int64(1))`. Two recipes
still show the uppercase `AVG(...)` / `COUNT(...)` form that was captured when
they were first written:

- `databricks` — 2 occurrences (`AVG(my_table.total_amount)`, `AVG(my_table.tip_amount)`, `COUNT(Int64(1))`)
- `refresh-data-window` — 2 occurrences (`COUNT(Int64(1))`)

The cookbook already contradicts itself here: `s3/README.md` and
`dremio/README.md` run the *identical* aggregate against the same `taxi_trips`
data and show the lowercase form, with the same row values. This PR brings the
two stale recipes in line with the runtime and with the rest of the cookbook.

Column widths are unchanged (`AVG`→`avg`, `COUNT`→`count`), so the ASCII table
borders still line up.

## Verified against

spiceai/spiceai at `v2.1.4` — ran the query on current stable
(`Runtime v2.1.4+models.metal`) against `s3://spiceai-demo-datasets/taxi_trips/2024/`.

## Evidence

```
$ spice sql
> select avg(total_amount), avg(tip_amount), count(1), passenger_count from taxi_trips group by passenger_count order by passenger_count asc;
+------------------------------+----------------------------+-----------------+-----------------+
| avg(taxi_trips.total_amount) | avg(taxi_trips.tip_amount) | count(Int64(1)) | passenger_count |
|            float64           |           float64          |      int64      |      int64      |
+------------------------------+----------------------------+-----------------+-----------------+
| 25.327816939455715           | 3.07225997139681           | 31465           | 0               |
| 26.20523044547591            | 3.3712622884679444         | 2188739         | 1               |
| 29.520659930928016           | 3.717130211328879          | 405103          | 2               |
| 29.13830904429006            | 3.5370455392168125         | 91262           | 3               |
```

The per-group counts (31465, 2188739, 405103, 91262, …) match both recipes
exactly — only the header case had drifted.

For `refresh-data-window` the whole recipe was run end-to-end and is otherwise
accurate: the baseline loads 2,964,624 rows, adding
`refresh_data_window: 35040h` evicts exactly the 5 documented pre-2023 rows
leaving 2,964,619, and the earliest remaining row is the documented
`2023-12-31T23:39:17` trip. The `databricks` change is static — it needs
Databricks credentials to run — but the header format is produced by DataFusion,
not the connector, so it is the same string the `s3`/`dremio` recipes already show.


grokspice:
@copilot review

lukekim (approved):

PR #559, opened 2026-07-30
Fix hashed_partitioning recipe: correct copy-pasted Spicepod name

claudespice:
## Summary

`hashed_partitioning/spicepod.yaml` declares `name: duckdb-acceleration-qs` — a leftover copied from the `duckdb/accelerator` recipe, which builds its app with `spice init duckdb-acceleration-qs`.

- **What the recipe said:** Step 2 presents the configuration as `name: hashed-partitioning`.
- **What actually loads:** Step 1 is `git clone` + `cd cookbook/hashed_partitioning`, and Step 3 is `spice run` — so the reader runs the *shipped* file, which announces itself as `duckdb-acceleration-qs`.
- **What changed:** the shipped pod name now matches the README. One line; no dataset, connector, or acceleration setting was touched.

The two files were otherwise byte-identical, so this was the only divergence between what the README shows and what the recipe runs.

## Verified against

`spiceai/spiceai` at `v2.1.2`. The pod `name` is a free-form label (`crates/spicepod`), so this is a naming/consistency fix rather than a runtime failure — the recipe's queries and partition-pruning plans are unaffected. The colliding name is legitimately owned by [`duckdb/accelerator/README.md:18`](https://github.com/spiceai/cookbook/blob/trunk/duckdb/accelerator/README.md#L18).

## Evidence

Static review only — confirmed by diffing the README's Step 2 YAML block against the committed `spicepod.yaml`; the pod name was the sole difference.

PR #574, opened 2026-08-07
Fix mismatched Spicepod names in sqlite and cdc-debezium/sasl-scram recipes

claudespice:
## Summary

Both recipes are clone-and-run (`git clone … && cd cookbook/<recipe>`), so the
YAML in the README is meant to show the `spicepod.yaml` the reader already has
checked out. In each case the `name:` differs from the shipped file, so a reader
comparing the two sees a config that doesn't look like theirs:

| Recipe | Shipped `spicepod.yaml` | README YAML |
| --- | --- | --- |
| `sqlite/accelerator` | `name: sqlite` | `name: spice_app` (Before *and* After blocks) |
| `cdc-debezium/sasl-scram` | `name: cdc-debezium` | `name: cdc-debezium-sasl-scram` |

In `sqlite/accelerator` the README is the stale side — Step 3 says "Use text
editor to open `spicepod.yaml`" and then shows a `spice_app` placeholder, so the
README was corrected.

In `cdc-debezium/sasl-scram` the shipped file is the stale side — it carries
`name: cdc-debezium`, copy-pasted from the sibling PLAINTEXT recipe at
`cdc-debezium/spicepod.yaml`, while its own README and directory both use the
descriptive `cdc-debezium-sasl-scram`. So the spicepod was corrected.

Either way the descriptive name wins, which matches the fix direction taken in
#559 for the same defect in `hashed_partitioning`. After this change each
README's YAML block matches its shipped `spicepod.yaml` exactly, name and params
included.

## Verified against

spiceai/spiceai at `v2.1.4`. The `name:` field is the Spicepod's own identifier
and is not referenced anywhere else in either recipe (checked the READMEs,
`Makefile`, `docker-compose.yaml`, and `.env` files), so this is a
documentation-consistency fix with no behavioral effect.

## Evidence

`sqlite/accelerator` was run end-to-end on current stable and is otherwise
accurate — the shipped `name: sqlite` is what the runtime loads, the documented
`trip_distance`/`total_amount` rows match exactly, and enabling the accelerator
behaves as described:

```
2026-08-07T12:11:33.046776Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), results cache enabled.
2026-08-07T12:12:02.025349Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (sqlite:file), results cache enabled.
2026-08-07T12:12:08.723187Z  INFO runtime::accelerated_table::refresh_task: Loaded 2,964,624 rows (399.38 MiB) for dataset taxi_trips in 6s 696ms.
```

Query before acceleration `3.41s` → after `0.29s`, with the documented values
(`312722.3 / 22.15`, `97793.92 / 36.31`, …) matching row for row.

`cdc-debezium/sasl-scram` is static review only — it needs Docker plus a Kafka +
Debezium + MySQL stack — but the mismatch is visible by reading the two files.

### DIFF
--- a/cdc-debezium/sasl-scram/spicepod.yaml
+++ b/cdc-debezium/sasl-scram/spicepod.yaml
@@ -1,6 +1,6 @@
 version: v1
 kind: Spicepod
-name: cdc-debezium
+name: cdc-debezium-sasl-scram
 datasets:
   - from: debezium:cdc.inventory.customer_addresses
     name: cdc
--- a/sqlite/accelerator/README.md
+++ b/sqlite/accelerator/README.md
@@ -80,7 +80,7 @@ Before:
 ```yaml
 version: v1
 kind: Spicepod
-name: spice_app
+name: sqlite
 datasets:
   - from: s3://spiceai-demo-datasets/taxi_trips/2024/
     name: taxi_trips
@@ -98,7 +98,7 @@ After:
 ```yaml
 version: v1
 kind: Spicepod
-name: spice_app
+name: sqlite
 datasets:
   - from: s3://spiceai-demo-datasets/taxi_trips/2024/
     name: taxi_trips


grokspice:
@copilot review

lukekim (approved):

PR #591, opened 2026-08-18
feat(authorization): add Cedar authorization recipe

Jeadie:
## What

A new self-contained recipe, `authorization/`, that secures datasets with [Cedar](https://www.cedarpolicy.com/) authorization policies and OIDC/JWT identity. It shows four distinct access-control patterns from a single Spicepod:

| Pattern | Rule | Demonstrated by |
| --- | --- | --- |
| **Multi-tenancy** | `tenant_id = current_org_id()` | Acme and Globex users see different rows of `customers` |
| **Row-level security (RLS)** | `owner = current_user_id()` | Two users in the same tenant see different rows of `deals`; an `admin` sees all |
| **PII masking** | `@mask_email` / `@mask_ssn`, by role | Analysts see masked PII; a PII officer sees real values |
| **RBAC** | Role-gated dataset + deny-by-default | Only the `hr` role may read `salaries` |

Multi-tenancy isolates by organization; RLS isolates by user — a finer grain the tenant filter alone cannot express. The `admin` role is exempt from the ownership filter and granted an unfiltered read, so a manager sees every deal.

## Identity

To keep the recipe self-contained, the identity provider is a small static file server (`serve-jwks.py`, standard library only) that serves an OIDC discovery document and a public JWKS. Four demo tokens are pre-signed and committed, so the recipe runs with no setup. In production you point `issuer_url` at your real identity provider and change nothing else.

The demo signing key is **not** committed — it is generated on demand by `generate-tokens.js` and gitignored. It is not needed to run: Spice validates the pre-signed tokens against the committed public JWKS.

## Notes

- **Requires the Enterprise distribution of Spice.ai** — Cedar authorization policies and OIDC identity are enterprise features. The README states this up front.
- **Unregistered from CI** — the recipe needs enterprise auth features, so it is intentionally not added to the `spice-qa.yml` matrix.
- All data is synthetic (fake tenants `acme`/`globex`, `example` emails, placeholder SSNs).

## Test

Verified end-to-end against an enterprise `spiced`: each README command was run and its output matches the committed expected output — tenant isolation, per-user RLS with the admin override, role-conditional masking, and deny-by-default RBAC all behave as documented.


grokspice:
**Pushed `114d4b5` to clear the `CodeQL` failure that has been blocking this PR.** It was the only red check; nothing else here needed changing, and I have not touched the recipe itself.

### What was failing

`CodeQL` reported *"1 new alert including 1 high severity security vulnerability"* — `js/file-system-race` at `authorization/generate-tokens.js:45`, the key write. The script asked whether `keys/demo-private-key.pem` existed and then wrote it in the `else`, so anything that created the key between those two steps was silently overwritten.

That matters here beyond the generic TOCTOU shape, because of how this recipe is put together: the `.pem` is gitignored while `jwks/jwks.json` and `tokens.env` are **committed, already signed by that key**. A clobbered key is not a lost throwaway file — it desynchronises the committed public JWKS from the committed tokens, and the recipe then fails validation for a reason the reader has no way to see.

### The fix

Read first and treat `ENOENT` as "not present", instead of testing and then writing:

- The check-then-act pair is gone entirely, so the alert has no source to flow from — this clears the rule rather than suppressing it.
- The create uses `flag: "wx"` (`O_CREAT|O_EXCL`), so a key that appeared in the meantime makes the write **fail** instead of replacing it. `O_EXCL` also refuses to follow a symlink at the final path component, which is the variant of this race actually worth defending against.
- On `EEXIST` the existing key is adopted, so the JWKS this run writes still matches the tokens it writes. Losing that would have traded one desync for another.
- `mode: 0o600` — a private key file should not be created at the umask default. Same call, and it is the one line in the diff that is an improvement rather than a fix; say so if you would rather it came out.

Behaviour is otherwise unchanged: an existing key is still reused, a missing one is still minted, and every non-`ENOENT` read error and non-`EEXIST` write error is rethrown rather than swallowed. The `const { privateKey }` destructuring stays inside the block, so it still shadows the outer `const privateKey` exactly as before.

### What I verified, and what I did not

- The diff is 16 insertions / 5 deletions in one file, identical under `--ignore-all-space`, no line-ending change.
- No `existsSync`/`stat`/`access` call remains in the file, so there is no second instance of this pattern.

**I could not execute it.** There is no Node on this host, so the change is verified by reading rather than by running `node generate-tokens.js`. The `CodeQL` `javascript-typescript` job parses the file, so a syntax error will surface as a red check rather than quietly — worth watching this run land before assuming it is green. An adversarial review pass was attempted and is a declared skip: the `codex` engine reported *"Your workspace is out of credits."*

This PR has been open 10 days and is otherwise `MERGEABLE`; once `CodeQL` is green it needs a review, not more CI.

lukekim (approved):

PR #610, opened 2026-08-24
Fix hashed_partitioning recipe: correct the Cayenne EXPLAIN transcripts

claudespice:
## Summary

Step 4 of the Hashed Partitioning recipe is wrong in two ways on current stable (v2.1.5):

1. **Both plan transcripts are from a DuckDB-era runtime.** The recipe accelerates with `engine: cayenne`, but the pasted plans show `DuckSqlExec sql= SELECT "VendorID", ... FROM "taxi_trips"` under `CooperativeExec`, and the prose tells the reader they will see *"the 10 `DuckSqlExec` plans"*. A Cayenne acceleration actually plans as `CayenneAccelerationExec: snapshots_scanned=1, files_scanned=1` over a `DataSourceExec ... file_type=vortex` reading one `expr0=<bucket>` directory per partition.

2. **The filtered query in the recipe errors out.** `EXPLAIN SELECT * FROM taxi_trips WHERE PULocationID = 221;` — the query that is supposed to demonstrate pruning — returns a `physical_plan_error` instead of a plan, and running the same filter without `EXPLAIN` fails outright:

```
sql> SELECT * FROM taxi_trips WHERE PULocationID = 221 LIMIT 2;
Internal Error: Unexpected internal error. Use '.error' for details.

$ curl -s -X POST http://localhost:8090/v1/sql -d "SELECT store_and_fwd_flag FROM taxi_trips WHERE PULocationID = 221"
Failed to execute query: Internal error: Assertion failed: lower.data_type() == upper.data_type()
(left: Utf8, right: LargeUtf8): Endpoints of an Interval should have the same type.
```

That is a runtime bug rather than a recipe bug (details below), so this PR keeps the recipe runnable by demonstrating pruning with `EXPLAIN SELECT COUNT(*) FROM taxi_trips WHERE PULocationID = 221;`, which projects only the partitioned column, prunes to a single partition, and matches how the sibling [`acceleration/partitioning`](https://github.com/spiceai/cookbook/tree/trunk/acceleration/partitioning) recipe demonstrates the same thing.

Both transcripts in this PR are captured verbatim from a live v2.1.5 run of this spicepod; only the Vortex file paths are abbreviated, and the README says so.

## The underlying runtime bug (worth a separate issue)

Any filtered query that projects a `LargeUtf8` column from a Cayenne acceleration fails with the interval assertion above. Reproduced on v2.1.5 with the dataset this recipe uses:

| acceleration | `SELECT store_and_fwd_flag FROM t WHERE PULocationID = 221` |
| --- | --- |
| none (S3 parquet source) | works |
| `engine: arrow` | works |
| `engine: duckdb` | works |
| `engine: cayenne, mode: file` | **Internal error** |

It is not partition-specific — an unpartitioned Cayenne acceleration of the same S3 parquet dataset fails identically — and it needs a source whose string columns are `LargeUtf8`: a Cayenne acceleration of a local CSV, whose columns are typed `Utf8`, is fine.

The cause looks like a stats/schema type mismatch: `crates/cayenne/src/stats.rs` maps a Vortex `DType::Utf8` min/max to `ScalarValue::Utf8` in `scalar_to_df`, while the accelerated table exposes the column as `LargeUtf8` (`describe taxi_trips` shows `store_and_fwd_flag | LargeUtf8`), so the two endpoints of the pruning interval disagree. That code path is unchanged on `trunk`.

## Verified against

spiceai/spiceai at v2.1.5 (current stable; CLI and runtime `v2.1.5+models.metal`), plus `origin/trunk` for the stats code path.

## Evidence

Live run of `hashed_partitioning/spicepod.yaml`:

```
Loaded 2,964,624 rows (399.38 MiB) for dataset taxi_trips in 3s 24ms.
All components are loaded. Spice runtime is ready!
```

`EXPLAIN SELECT COUNT(*) FROM taxi_trips WHERE PULocationID = 221;` prunes to one partition:

```
|               |               CayenneAccelerationExec: snapshots_scanned=1, files_scanned=1
|               |                 BytesProcessedExec
|               |                   DataSourceExec: file_groups={1 group: [[.spice/data/taxi_trips/expr0=7/.../..._00000.vortex]]}, projection=[PULocationID], file_type=vortex, predicate: PULocationID@7 = 221
```

and the unfiltered `EXPLAIN SELECT * FROM taxi_trips;` shows all ten partitions, `expr0=0` through `expr0=9`.


claudespice:
@copilot review

lukekim (approved):

PR #611, opened 2026-08-24
Fix cayenne recipe: refresh the startup and query transcripts

claudespice:
## Summary

Ran the Cayenne recipe end to end on current stable (v2.1.5) and every transcript in it was stale:

| Recipe says | v2.1.5 actually prints |
| --- | --- |
| `Loaded 2,964,624 rows (421.71 MiB) ... in 16s 167ms` | `Loaded 2,964,624 rows (399.38 MiB) ... in 10s 302ms` |
| cache line *after* the Flight/HTTP listener lines | cache line *before* them, and `Starting runtime v2.1.5+models.metal` above both |
| `Dataset taxi_trips registered (...), results cache enabled.` | same line plus `duration_ms=18` |
| query tables with a header row only | header row plus the data-type row (`int32`, `timestamp[us]`, `int64`) that v2 `spice sql` prints |

The 421.71 MiB figure is the same pre-v2 accelerated size that #588 corrected in the sibling [`acceleration/partitioning`](https://github.com/spiceai/cookbook/tree/trunk/acceleration/partitioning) recipe; this recipe accelerates the identical dataset with the identical engine, so it drifted the same way.

Row values in Steps 4 and 7 also differ from what the recipe shows — the `LIMIT 10` has no `ORDER BY`, so the rows the runtime returns are whatever the scan yields first (they differed between my own two runs). The tables here are the verbatim output of one live run rather than a promise of specific rows.

Config, steps, version floor (`v1.9+`), and links all check out unchanged.

## Verified against

spiceai/spiceai at v2.1.5 (current stable), CLI and runtime `v2.1.5+models.metal`.

## Evidence

Step 6, accelerated (`engine: cayenne`, `mode: file`):

```
2026-08-24T20:58:43.033807Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (cayenne:file), results cache enabled. duration_ms=477
2026-08-24T20:58:43.035086Z  INFO runtime::accelerated_table::refresh_task: Loading data for dataset taxi_trips
2026-08-24T20:58:53.353320Z  INFO runtime::accelerated_table::refresh_task: Loaded 2,964,624 rows (399.38 MiB) for dataset taxi_trips in 10s 302ms.
2026-08-24T20:58:53.400890Z  INFO runtime: All components are loaded. Spice runtime is ready!
```

The query the recipe uses took `1.256426375 seconds` against S3 and `0.064644125 seconds` once accelerated, so the point of the recipe still lands.

### DIFF
--- a/cayenne/README.md
+++ b/cayenne/README.md
@@ -41,10 +41,12 @@ Confirm in the terminal output the `taxi_trips` dataset has been loaded:
 
 ```bash
 Spice.ai runtime starting...
-2024-09-16T21:25:43.306009Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
-2024-09-16T21:25:43.309474Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:8090
-2024-09-16T21:25:43.507974Z  INFO runtime::init::caching: Initialized sql results cache; max size: 128.00 MiB, item ttl: 1s, hashing algorithm: XXH3, encoding: none
-2024-09-16T21:25:44.101055Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), results cache enabled.
+2026-08-24T20:57:43.603253Z  INFO spiced: Starting runtime v2.1.5+models.metal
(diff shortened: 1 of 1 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #619, opened 2026-09-05
Fix search recipes: document /v1/search matches as an array

claudespice:
## Summary

Four recipes document the `POST /v1/search` response with `matches` holding a bare string:

```json
"matches": {
  "content": "# Metrics Naming\n\n## TL;DR\n..."
}
```

The runtime returns a list. `Match.matches` is `HashMap<String, Vec<Value>>` — a single column can contribute more than one highlight to a row (a chunked column commonly does) — so the value is always a JSON array, even when it holds one element:

```json
"matches": {
  "content": ["# Metrics Naming\n\n## TL;DR\n..."]
}
```

This changes `azure_openai`, `models/openai`, `vectors/s3` and `search_github_files` (16 blocks). Only the shape changes — every highlight string, score, and data value is left exactly as it was. The `full-text-search` recipe already documented the array form, which is what made the inconsistency visible.

## Verified against

spiceai/spiceai at `v2.2.1` — [`crates/runtime-search/src/types.rs`](https://github.com/spiceai/spiceai/blob/v2.2.1/crates/runtime-search/src/types.rs):

```rust
pub struct Match {
    /// The matches for this result
    matches: HashMap<String, Vec<Value>>,
    ...
}
```

`git show` on the `v2.0.0`, `v2.1.0`, `v2.2.0` and `v2.2.1` tags gives the same declaration, so this shape is not new in `v2.2.1` — the samples have been wrong for every `v2.x` reader.

## Evidence

Runtime output from a local `spiced v2.2.1+models.metal` (a `file:` dataset with an `all-MiniLM-L6-v2` embedding on one text column, no chunking — the simplest case, which still returns an array):

```console
$ curl -s -X POST http://localhost:8090/v1/search -H 'Content-Type: application/json' \
    -d '{"text":"animals","additional_columns":["title"],"limit":1}'
{
    "results": [
        {
            "matches": {
                "description": [
                    "A Touching Reflection of a Crocodile And a Dog who must Chase a Hunter in An Abandoned Fun House"
                ]
            },
            "data": { "title": "STRICTLY SCARFACE" },
            "_score": 0.7381305772506472,
            "dataset": "spice.public.films"
        }
    ],
    "duration_ms": 173
}
```

Note for reviewers: the upstream OpenAPI example in [`crates/runtime/src/http/v1/search.rs`](https://github.com/spiceai/spiceai/blob/v2.2.1/crates/runtime/src/http/v1/search.rs) still shows the scalar form too, which is likely where these samples came from. That one is worth a separate fix in `spiceai/spiceai`.

### DIFF
--- a/azure_openai/README.md
+++ b/azure_openai/README.md
@@ -137,7 +137,9 @@ Result:
   "results": [
     {
       "matches": {
-        "content": ".\n\n## Definitions\n\n- Metric: is a measurement used to track the state and behavior of a system component. Metrics represent the current status ..."
+        "content": [
+          ".\n\n## Definitions\n\n- Metric: is a measurement used to track the state and behavior of a system component. Metrics represent the current status ..."
+        ]
       },
       "data": {
         "download_url": "https://raw.githubusercontent.com/spiceai/spiceai/trunk/docs/dev/metrics.md"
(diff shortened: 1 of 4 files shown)


claudespice:
@copilot review

lukekim (approved):

PR #621, opened 2026-09-06
Fix gospice SDK sample: cloud snippet is not where the README says

claudespice:
## Summary

`client-sdk/gospice-sdk-sample` documents the Spice.ai Cloud snippet as `cloud/main.go` and tells the reader to run it with `go run ./cloud`:

> - `cloud/main.go`: Query Spice.ai Cloud with inline replacement values.
>
> The cloud snippet keeps an inline API key placeholder by design. Replace the API key placeholder in `cloud/main.go` with `${SPICE_API_KEY}`, then run:
>
> ```bash
> go run ./cloud
> ```

There is no `cloud/` directory. The file ships as `main_cloud.go` in the package root, so:

- `go run ./cloud` fails — the directory does not exist.
- `go build ./...` / `go vet ./...` fail — `main_cloud.go` declares a second `func main()` in `package main`, which is a redeclaration. (The documented local flow, `go run main.go`, happens to work because it compiles only the named file.)

The README has pointed at `cloud/main.go` since #369; the file was never moved.

This moves `main_cloud.go` → `cloud/main.go`, which both matches the README and gives the snippet its own package so the module compiles as a whole. Two smaller fixes in the same README:

- Prerequisite `Go 1.24+` → `Go 1.25+`; `go.mod` declares `go 1.25.0`.
- Links section pointed at `pkg.go.dev/github.com/spiceai/gospice/v6` while the sample imports `github.com/spiceai/gospice/v8`.

No sample code changed — the moved file is byte-identical.

Every other SDK sample already names its real cloud file (`main_cloud.py`, `index_cloud.mjs`, `Cloud.cs`, `Cloud.java`), which is what made this one stand out.

## Verified against

Spice runtime `v2.2.1` (current stable) and Go 1.27.1, `gospice/v8 v8.0.1` as pinned in `go.mod`.

## Evidence

Before, on a clean checkout of `trunk`:

```console
$ go run ./cloud
stat .../gospice-sdk-sample/cloud: directory not found

$ go build ./...
# gospice-sdk-sample
./main_cloud.go:10:6: main redeclared in this block
	./main.go:11:6: other declaration of main
```

After this change:

```console
$ go build ./... && echo ok
ok
$ go vet ./... && echo ok
ok
$ go run ./cloud
panic: Error initializing client: apiKey is invalid
```

`go run ./cloud` now compiles and reaches the runtime, failing only on the deliberate `"API_KEY"` placeholder — which is the point where the reader substitutes their own key.

The documented local flow still works unchanged against a live `spiced v2.2.1+models.metal` with this sample's spicepod:

```console
$ go run main.go
=== Using Sql ===
VendorID: 2, tpep_pickup_datetime: 1706687247000000, fare_amount: 9.3
VendorID: 2, tpep_pickup_datetime: 1706685623000000, fare_amount: 12.1
...

=== Using SqlWithParams ===
VendorID: 2, tpep_pickup_datetime: 1706687247000000, fare_amount: 9.3
...
```


claudespice:
@copilot review

lukekim (approved):

PR #628, opened 2026-09-07
Fix mysql/planetscale and postgres/rds: READMEs name placeholders the Spicepods don't contain

claudespice:
## Summary

Both recipes are templates: the reader edits bracketed placeholders in `spicepod.yaml` before
running. In two of them the README names placeholders that are not in the file, so following
the step means searching for a token that isn't there.

**`mysql/planetscale`** — Step 4 says to replace `[mysql_db]` and `[mysql_user]`; the Spicepod
contains `[db-name]` and `[username]`. Renamed the two placeholders in the Spicepod, which also
makes them consistent with the `[mysql_host]` placeholder already sitting next to them.

**`postgres/rds`** — Step 3 says to replace `[pg_host]`, `[pg_port]`, `[pg_user]` and `[pg_db]`;
the Spicepod contains `[Host]`, `[Port]`, `[User]` and `[Database name]` (`pg_host` etc. are the
*param keys*, which stay). Reworded the step to name each param and the placeholder value it
holds. Fixed on the README side here so `postgres/rds` keeps the same placeholder style as its
siblings `postgres/connector` and `postgres/accelerator`.

## Verified against

The recipes themselves — this is an internal-consistency defect, provable from the two files:

```console
$ grep -oE '\[[A-Za-z_][^]]*\]' mysql/planetscale/spicepod.yaml | sort -u
[db-name]
[local_table_name]
[mysql_host]
[remote_table_path]
[username]

$ grep -oE '\[mysql_[a-z]*\]' mysql/planetscale/README.md | sort -u
[mysql_db]
[mysql_host]
[mysql_user]
```

Same shape for `postgres/rds`: the README's `[pg_db]` / `[pg_host]` / `[pg_port]` / `[pg_user]`
appear nowhere in `postgres/rds/spicepod.yaml`.

## Evidence

Static review, plus a v2.2.1 runtime run that shows why the placeholders matter: an unreplaced
bracket placeholder is parsed as a YAML flow sequence, so the Spicepod fails to load rather than
failing to connect —

```text
WARN spiced: ... Unable to load spicepod .../mysql/planetscale: Failed to parse spicepod.yaml:
Invalid type sequence. Expected a string
```

That behaviour is the existing convention across the template recipes and is not changed here;
only the placeholder names are aligned. No functional config was touched — after substitution
both recipes behave exactly as before.

Checked and left alone: `postgres/connector` and `postgres/accelerator` use the same
`[Host]`/`[Port]`/`[User]`/`[Database name]` placeholders but never name them in their READMEs,
so there is nothing inconsistent to fix there.


claudespice:
@copilot review

lukekim (approved):

PR #629, opened 2026-09-07
Refresh SQL transcripts for the v2 typed header (file, s3, sqlite/accelerator, duckdb/accelerator)

claudespice:
## Summary

Since v2.0 the SQL REPL renders results through `format_batches_with_types`, so every real
`spice sql` table carries a data-type row under centered column names. These four recipes still
showed the pre-v2 untyped, left-aligned header, so the table a reader sees on screen does not
match the one in the README.

Eight tables across four recipes, each type taken from running that recipe's own query on
v2.2.1 — not inferred. The recipes' captured values are left as they were; only the header
changes, plus one column width in `s3` (`table_name` was padded a character wide).

| Recipe | Types confirmed by running |
| --- | --- |
| `file` | `float64` (avg over `passenger_count`), `varchar` (`location`) |
| `s3` | 4× `varchar` (`show tables`), `float64`/`float64`/`int64`/`int64` (the group-by) |
| `sqlite/accelerator` | `float64`/`float64` |
| `duckdb/accelerator` | `int32`/`timestamp[us]`/`timestamp[us]`/`int64` |

## Verified against

spiceai/spiceai at `v2.2.1` — `crates/repl/src/pretty.rs`, `format_batches_with_types`,
unconditional in non-expanded mode. Runtime and CLI v2.2.1 (`v2.2.1+models.metal`), macOS arm64.

## Evidence

Every recipe was run and every query executed as written.

`file` — downloaded `yellow_tripdata_2024-01.parquet` per Step 1, and the value in the README
reproduces exactly:

```console
$ echo 'select avg(passenger_count) from yellow_taxis;' | spice sql
+-----------------------------------+
| avg(yellow_taxis.passenger_count) |
|              float64              |
+-----------------------------------+
| 1.3392808966805005                |
+-----------------------------------+
```

`s3` — `s3://spiceai-demo-datasets/taxi_trips/2024/` loaded `2,964,624 rows (399.38 MiB)`,
matching the README's Step 2 log line. The `show tables` block in this PR is byte-identical to
the live output (`diff` clean), which is where the corrected `table_name` width comes from.

`sqlite/accelerator` — `SELECT trip_distance, total_amount FROM taxi_trips ORDER BY trip_distance
DESC LIMIT 10` returned exactly the ten rows already in the README, so only the type row is new.

`duckdb/accelerator` — the query has no `ORDER BY`, so its ten rows differ from run to run;
the recipe's own captured rows are kept and only the verified type row is added.

The `Time: … seconds. N rows.` trailer was left as each recipe had it — this PR does not change
that convention either way.

This is a first batch. 54 other recipes still carry pre-v2 untyped tables; the remainder need
credentials, Docker, or an Enterprise runtime to reproduce honestly, so they are better done as
their owners' recipes get run rather than by guessing types.


claudespice:
@copilot review

lukekim (approved):

PR #636, opened 2026-09-09
Fix 17 recipes: 'Checking for latest Spice runtime release' is not part of spice run output

claudespice:
## Summary

17 recipes open their `spice run` transcript with two lines that the v2 CLI never prints on that path:

```
2025/07/07 11:32:50 INFO Checking for latest Spice runtime release...
2025/07/07 11:32:50 INFO Spice.ai runtime starting...
```

Two problems:

1. **`Checking for latest Spice runtime release...` belongs to `spice install`.** `spice run` only reaches it when the runtime is *not installed*, and in that case it is preceded by `Spice.ai runtime is not installed. Installing now...` — which none of these transcripts show. Every one of them goes straight on to `Starting runtime`, so the runtime was already present and the line never appeared.
2. **The `2025/07/07 11:32:50 INFO` prefix is a v1 artifact.** That Go-style date/time prefix predates the Rust CLI, which emits the line with no timestamp at all.

Replaced both lines with the single line the CLI actually prints:

```
```

`graphql/` carries the same defect but is being fixed separately along with the rest of its (v1.5.0-era) transcript.

Note for reviewers: in an interactive terminal the CLI also prints a `Spice.ai OSS CLI v2.2.1` banner to **stderr** before this line. These transcripts are excerpts and most don't show it, so this PR doesn't add it — `spark/README.md` already has one (at a stale version) and it's left as-is.

## Verified against

spiceai/spiceai at `v2.2.1`:

- `bin/spice/src/commands/install.rs:94` — `tracing::info!("Checking for latest Spice runtime release...")`, inside `install::execute`.
- `bin/spice/src/runtime_launcher.rs:137-141` — `spice run` calls `install::execute` only under `if !ctx.is_runtime_installed()`, logging `Spice.ai runtime is not installed. Installing now...` first.
- `bin/spice/src/runtime_launcher.rs:150` — `tracing::info!("Spice.ai runtime starting...")`, unconditional.
- `bin/spice/src/main.rs:342` — the `Spice.ai OSS CLI <version>` banner, stderr and TTY only.

## Evidence

`spice run` against a local `spiced v2.2.1+models.metal` with the runtime already installed:

```
$ head -2 out.log
2026-09-09T12:19:15.766737Z  INFO spiced: Starting runtime v2.2.1+models.metal
```

`od -c` on the first line confirms the exact prefix is a single leading space, no timestamp:

```
0000000        I   N   F   O       S   p   i   c   e   .   a   i       r
0000020    u   n   t   i   m   e       s   t   a   r   t   i   n   g   .
0000040    .   .  \n
```

No occurrence of `Checking for latest Spice runtime release` appeared in that output.


claudespice:
@copilot review

lukekim (approved):

PR #651, opened 2026-09-13
Fix refresh-data-window: SQL transcripts predate the v2 typed header, and the first log block is from 2024

claudespice:
## Summary

Two stale transcripts in `refresh-data-window/README.md`, both refreshed from a real run of the recipe's own spicepod on the current stable runtime.

1. **The four `spice sql` result blocks show the pre-v2 table header.** Since v2.0 the REPL prints a data-type row under the column names, and that row widens any column whose type name is longer than its name — `extra` becomes `float64`, so the separator rows differ too. A reader comparing their output to the README sees a different-looking table on every query.
2. **The Step 3 runtime log block was captured in 2024** (`2024-08-05T14:35:55`). It is missing `duration_ms=0` on the registration line, prints Flight/HTTP *before* the dataset starts initializing, and omits the startup and ready lines. Step 4's block was refreshed in #632, so the two halves of one transcript were three weeks and two years apart; both now come from the same run.

Nothing else in the recipe changed: the row counts (2,964,624 → 2,964,619), the five evicted rows, the earliest remaining pickup (`2023-12-31T23:39:17`), and the accelerated size (399.38 MiB) are all still exactly what the README claims.

## Verified against

spiceai/spiceai at `v2.3.0` (current stable, `v2.3.0+models.metal`), run from an empty directory with the Step 2 spicepod copied verbatim from this README.

## Evidence

**Reproduced.** Step 3, before (README) vs the run:

```console
$ spice sql
+-----------------+
| count(Int64(1)) |
|      int64      |      <- not in the README
+-----------------+
| 2964624         |
+-----------------+

Time: 0.001221792 seconds. 1 rows.
```

Startup log from the same run — compare with the 2024 block the README shipped:

```console
2026-09-13T12:09:52.083872Z  INFO spiced: Starting runtime v2.3.0+models.metal
2026-09-13T12:09:52.088606Z  INFO runtime::init::dataset: Dataset taxi_trips initializing...
2026-09-13T12:09:52.288168Z  INFO runtime::flight: Spice Runtime Flight listening on 127.0.0.1:50051
2026-09-13T12:09:52.288514Z  INFO runtime::http: Spice Runtime HTTP listening on 127.0.0.1:8090
2026-09-13T12:09:53.154126Z  INFO runtime::init::dataset: Dataset taxi_trips registered (s3://spiceai-demo-datasets/taxi_trips/2024/), acceleration (arrow), results cache enabled. duration_ms=0
2026-09-13T12:09:53.155485Z  INFO runtime_table::accelerated::refresh_task: Loading data for dataset taxi_trips
2026-09-13T12:09:57.995471Z  INFO runtime_table::accelerated::refresh_task: Loaded 2,964,624 rows (399.38 MiB) for dataset taxi_trips in 4s 839ms.
2026-09-13T12:09:58.030404Z  INFO runtime: All components are loaded. Spice runtime is ready!
```

Then adding `refresh_data_window: 35040h` to the running spicepod reloaded it to 2,964,619 rows and `select * ... limit 1` returned the `2023-12-31T23:39:17` trip — the recipe's claim of "5 rows difference" still holds today.


claudespice:
@copilot review

lukekim (approved):

PR #652, opened 2026-09-13
Fix caching/accelerator: document the two warnings the shipped config still prints

claudespice:
## Summary

Ran this recipe end-to-end on v2.3.0. The caching behaviour it documents is exactly right — SWR, the cache-key-per-path behaviour, and all eight columns of the schema table match. What does not match is the startup output: the spicepod as shipped prints two `WARN`s that the README never mentions, and #642 (which added `caching_max_size` / `caching_max_items` to silence a third) was written without a run, so a reader who follows it has no way to tell whether the budgets took effect.

- **`runtime::datafusion` — the retention warning.** The budgets bound how much is *stored*, not how *old* it gets, so the time-based bound is still absent and the runtime still says so. The README already explains this distinction in prose ("the runtime warns separately when none is running"); it just never shows the warning, so the reader sees a scary paragraph-long WARN about unbounded growth immediately after being told the budgets fix unbounded growth.
- **`runtime_table::accelerated` — `on_zero_results` is ignored in caching mode.** This one fires for *every* caching-mode dataset regardless of whether the key is set, and this spicepod never sets it — so its remediation, "Remove `on_zero_results` from the dataset configuration", cannot be followed here.

Also corrects the quoted no-budget warning to the string the runtime actually emits: an em dash rather than a hyphen, and the trailing `For details, visit:` link.

## Verified against

spiceai/spiceai at `v2.3.0` — [`crates/runtime-table/src/accelerated/mod.rs:1096`](https://github.com/spiceai/spiceai/blob/v2.3.0/crates/runtime-table/src/accelerated/mod.rs#L1096) (the `on_zero_results` warning, guarded only by `if refresh_mode == RefreshMode::Caching`), [`crates/runtime-table/src/accelerated/caching_eviction.rs:288`](https://github.com/spiceai/spiceai/blob/v2.3.0/crates/runtime-table/src/accelerated/caching_eviction.rs#L288) (the budget warning), and [`crates/runtime/src/datafusion/caching_retention.rs:138`](https://github.com/spiceai/spiceai/blob/v2.3.0/crates/runtime/src/datafusion/caching_retention.rs#L138) (the retention warning).

## Evidence

**Reproduced.** `spice run` with this recipe's `spicepod.yaml` unmodified, against the recipe's own time server on `:7400`:

```console
2026-09-13T12:12:31.696869Z  WARN runtime::datafusion: Dataset 'time' sets `caching_stale_if_error: enabled` and has no retention policy running, so no cached entry is ever evicted and the accelerator grows with every distinct request it serves. ...
2026-09-13T12:12:31.697994Z  WARN runtime_table::accelerated: Dataset time: `on_zero_results` is ignored when `refresh_mode: caching` is set. Caching mode always queries the source on a cache miss. Remove `on_zero_results` from the dataset configuration to silence this warning. ...
2026-09-13T12:12:31.699479Z  INFO runtime::init::dataset: Dataset time registered (http://localhost:7400), acceleration (duckdb, caching). duration_ms=83
```

The same spicepod with the two budget lines deleted adds the third warning, which is where the corrected quote comes from:

```console
2026-09-13T12:14:52.707341Z  WARN runtime_table::accelerated::caching_eviction: Dataset 'time' sets `caching_stale_if_error: enabled` with no `caching_max_size` or `caching_max_items`, so no cached entry is ever evicted and the acceleration will grow without bound — expired entries are deliberately kept as fallback for a failing origin. Set a budget to bound it. For details, visit: https://spiceai.org/docs/components/data-accelerators/data-refresh#refresh-modes
```

Everything else in the recipe checked out on the same run and is left alone — first query ~1.02s, repeat queries ~0.002s, and SWR serving stale then catching up:

```console
12:13:31  | 2026-09-13T12:13:18Z | 2026-09-13T12:13:19.282296 |
12:13:43  | 2026-09-13T12:13:31Z | 2026-09-13T12:13:32.771427 |
12:13:55  | 2026-09-13T12:13:43Z | 2026-09-13T12:13:44.832702 |
```


claudespice:
@copilot review

lukekim (approved):

PR #653, opened 2026-09-13
Fix three documentation links whose anchors no longer exist

claudespice:
## Summary

Swept every fragment URL in the repo — 24 links carrying a `#anchor` — and checked that the anchor actually exists on the page it points to. Three do not. All three pages still return `200`, so the repo's usual status-code link check passes them; the reader just lands at the top of a long page and has to hunt for the section.

| Recipe | Link | What happens now |
| --- | --- | --- |
| `clickhouse` | `clickhouse.com/docs/en/install#quick-install` | Redirects to `/docs/get-started/setup/install`, dropping the fragment. That page's "Quick install" heading now carries a generated id (`_R_cillmtbsnlhjiuqsnpfiutb_`), so there is no stable anchor left to point at — link to the canonical `/docs/install` with no fragment. |
| `architectures/microservice` | `spiceai.org/docs/features/observability#metrics` | The page has no `metrics` id. The section the sentence means ("Spice provides several metrics…") is `#available-metrics`. |
| `llm-judge` | `spiceai.org/docs/features/large-language-models/evals#llm-judge` | The evals page is now a deprecation stub — evals were removed in v2.0 — with no content anchors at all. |

For `llm-judge` the fix is the versioned page rather than a different section: the recipe already says at the top that it needs a `v1.x` release, and the stub itself refers the reader to the v1.11.x evals documentation, which still has the `llm-judge` section.

## Verified against

The live docs; `spiceai/spiceai` at `v2.3.0` for the evals removal the stub describes (`spiceai/spiceai#9420`, removed in v2.0.0 — already noted in the recipe).

## Evidence

**Reproduced.** Anchor presence checked by fetching each page and grepping the rendered ids (Docusaurus emits unquoted `id=` attributes, so the check accounts for both spellings). Before:

```console
MISSING  https://clickhouse.com/docs/en/install#quick-install
MISSING  https://spiceai.org/docs/features/observability#metrics
MISSING  https://spiceai.org/docs/features/large-language-models/evals#llm-judge
OK       (the other 21)
```

After, on the three replacement URLs:

```console
200 (no anchor) https://clickhouse.com/docs/install
200 ANCHOR-OK   https://spiceai.org/docs/features/observability#available-metrics
200 ANCHOR-OK   https://spiceai.org/docs/v1.11/features/large-language-models/evals#llm-judge
```

The same sweep over in-repo `](#…)` links (8 of them) found every anchor resolving to a heading in its own file, so nothing there needed changing.


claudespice:
@copilot review

lukekim (approved):

PR #659, opened 2026-09-16
Fix search_github_files: full-text-search result is short a row, and no wait for embeddings

claudespice:
## Summary

Two defects found by running the recipe end-to-end on Spice `v2.3.1`.

**1. The full-text-search response is missing a result and a field.** The request sends `"limit": 3`, but the documented response shows two results and ends at `"results": [...]` with no `duration_ms` — while every other `/v1/search` response in the same README carries `duration_ms`. The runtime returns three results and `duration_ms`. A reader diffing their output against the README sees an extra result and an extra field and cannot tell which is right. The first result's `matches` keys were also transcribed in the opposite order from the response.

**2. Enabling embeddings makes startup take minutes, and the recipe never says so.** Step 2 is "Restart the `spiced` runtime" and step 3 is "Perform a basic search". Followed in order, the search runs while the acceleration is still loading and fails. Measured on this run: **5s** to ready with the embeddings block commented out, **3m 55s** and **3m 57s** (two runs) with it uncommented — the local `all-MiniLM-L6-v2` model chunks and embeds ~200 Markdown files.

## Verified against

spiceai/spiceai at `v2.3.1` (current stable, released 2026-09-14), runtime `v2.3.1+models.metal`, run against this recipe's own Spicepod with a GitHub token.

## Evidence

**Reproduced.**

Everything else in the recipe was confirmed unchanged on this run and is untouched by this PR: the SQL query returns the same 22 rows, and both vector-search examples and the child-runtime example return the documented results **including the exact `_score` values** (`0.7255329489486705`, `0.7108545813115704`, `0.7487837074615193`, `0.746478348538512`).

### 1. Full-text search — `"limit": 3` returns three results

Request exactly as the README gives it:

```console
$ curl -s -XPOST http://localhost:8090/v1/search -H "Content-Type: application/json" \
    -d '{"datasets": ["doc.pulls"], "text": "Glue data", "limit": 3}'
```

The third result the README omits, and the trailing field:

```json
    {
      "matches": {
        "title": [
          "Document Snowflake Data Connector"
        ],
        ...
      },
      "primary_key": {
        "id": "PR_kwDOF38K0s5uby1C"
      },
      "_score": 0.029910714285714284,
      "dataset": "doc.pulls"
    }
  ],
  "duration_ms": 7
}
```

### 2. Searching before the load finishes

With the embeddings block uncommented, a search 20s after restart:

```console
$ curl -s -XPOST http://localhost:8090/v1/search -H "Content-Type: application/json" \
    -d '{"datasets": ["spiceai.files"], "text": "testing", "limit": 1}'
Failed to execute search query: External error: Acceleration not ready; loading initial data for spiceai.files
```

Load time without the embeddings block:

```text
INFO runtime_table::accelerated::refresh_task: Loaded 207 rows (4.69 MiB) for dataset spiceai.files in 5s 128ms.
```

and with it, on two separate runs:

```text
INFO runtime_table::accelerated::refresh_task: Loaded 207 rows (12.75 MiB) for dataset spiceai.files in 3m 55s 719ms.
INFO runtime_table::accelerated::refresh_task: Loaded 207 rows (12.75 MiB) for dataset spiceai.files in 3m 57s 67ms.
```

Row counts and `duration_ms` track a live repository, so the note above the SQL results already tells readers those move; the third result is added because the response shape — three results for `limit: 3`, plus `duration_ms` — does not.


claudespice:
@copilot review

lukekim (approved):

PR #661, opened 2026-09-16
Fix views recipe: SQL transcripts predate the v2 typed result header

claudespice:
## Summary

All three query results in this recipe show the v1 rendering — left-aligned column names, no data-type row. Since v2.0 the SQL REPL centers the names and prints their types beneath.

Only the header rows change. The row data is untouched, because running the recipe reproduces every documented row exactly.

## Verified against

spiceai/spiceai at `v2.3.1` (current stable, released 2026-09-14), runtime `v2.3.1+models.metal`.

## Evidence

**Reproduced** — the recipe was run end-to-end as written: `spice init accelerated-views`, `spice add spiceai/tpch`, the view appended to `spicepod.yaml` verbatim from Step 3, then each of the three queries.

The view loaded with the size the recipe documents:

```text
INFO runtime::datafusion: View supplier_order_waits registered, acceleration (duckdb, 3600s refresh).
INFO runtime_table::accelerated::refresh_task: Loaded 10,000 rows (481.43 kiB) for view supplier_order_waits in 1s 22ms.
```

Step 4, first query — every documented row matches; only the header differs:

```console
$ echo "SELECT * FROM supplier_order_waits LIMIT 5;" | spice sql
+--------------------+------------+---------+
|       s_name       |   nation   | numwait |
|       varchar      |   varchar  |  int64  |
+--------------------+------------+---------+
| Supplier#000002340 | INDIA      | 25      |
| Supplier#000003591 | JAPAN      | 24      |
| Supplier#000000014 | MOROCCO    | 23      |
| Supplier#000001915 | JORDAN     | 23      |
| Supplier#000007754 | MOZAMBIQUE | 23      |
+--------------------+------------+---------+

Time: 0.002188917 seconds. 5 rows.
```

Q21 run directly, and then through the view — again, the ten documented rows match in both:

```console
+--------------------+--------------+---------+
|       s_name       |    nation    | numwait |
|       varchar      |    varchar   |  int64  |
+--------------------+--------------+---------+
| Supplier#000002829 | SAUDI ARABIA | 20      |
| Supplier#000005808 | SAUDI ARABIA | 18      |
| Supplier#000000262 | SAUDI ARABIA | 17      |
...
```

## Not addressed

The recipe's speedup figures (`0.259203917` → `0.005394292` seconds) still read as first written. This run measured `0.054553209` → `0.00217475` — the same conclusion, and both are one machine's numbers, so overwriting one set with another adds nothing. The `Time:` footers are left alone for the same reason.


claudespice:
@copilot review

lukekim (approved):

PR #666, opened 2026-09-19
Fix describe transcripts in five recipes: the v2 REPL prints a table_schema column

claudespice:
## Summary

Five recipes show a `describe` transcript in the pre-v2 shape. Three things are wrong with each, and all three are deterministic — they do not depend on the connector or the data:

1. **The `table_schema` column is missing.** The REPL rewrites a *bare* `describe <table>` into `select table_schema, column_name, data_type, is_nullable from information_schema.columns where table_name = '…'`. All five use the bare form, so the runtime returns four columns; the READMEs show three. (The three-column form is what a *two-part* `describe schema.table` returns — a different command from the one these recipes tell the reader to run.)
2. **The type row is missing.** Since v2 the REPL prints a `varchar` row under the header, which also widens the table's column rules.
3. **Timestamps use the pre-v2 spelling.** `kubernetes` and `spark` show `Timestamp(Microsecond, None)` / `Timestamp(Microsecond, Some("Etc/UTC"))`; v2.3.1 renders these `Timestamp(µs)` and `Timestamp(µs, "Etc/UTC")`.

Affected: `oracle`, `redshift`, `json_strings`, `kubernetes`, `spark`.

Every column name, Arrow type and `is_nullable` value in the transcripts is left untouched — only the table's shape, the added `table_schema` value, and the timestamp spelling changed. `table_schema` is `public` in each case because all five datasets are registered under the default schema, which each recipe's own `show tables` output in the same README already shows.

## Verified against

spiceai/spiceai at `v2.3.1` — `crates/repl/src/lib.rs:624-638`:

```rust
let rewritten = if let Some((schema, name)) = raw_ident.split_once('.') {
    format!("select column_name, data_type, is_nullable from information_schema.columns where table_schema = '{}' and table_name = '{}' order by ordinal_position;", …)
} else {
    format!("select table_schema, column_name, data_type, is_nullable from information_schema.columns where table_name = '{}' order by table_schema, ordinal_position;", …)
};
```

### Reproduction

**Reproduced** for `json_strings`, which is locally runnable (`file://tshirts.csv`), on the released v2.3.1 binary — `json_strings/README.md:94`.

On base (`577e486`), the README tells the reader to expect this:

```console
sql> describe products;
+-------------+-----------+-------------+
| column_name | data_type | is_nullable |
|   varchar   |  varchar  |   varchar   |
+-------------+-----------+-------------+
| id          | Int64     | YES         |
| name        | Utf8      | YES         |
| properties  | Utf8      | YES         |
+-------------+-----------+-------------+
```

Running that exact step prints a different table — four columns, not three:

```console
$ spice run
$ spice sql
sql> describe products;
+--------------+-------------+-----------+-------------+
| table_schema | column_name | data_type | is_nullable |
|    varchar   |   varchar   |  varchar  |   varchar   |
+--------------+-------------+-----------+-------------+
| public       | id          | Int64     | YES         |
| public       | name        | Utf8      | YES         |
| public       | properties  | Utf8      | YES         |
+--------------+-------------+-----------+-------------+

Time: 0.077833541 seconds. 3 rows.

sql> describe products_with_color;
+--------------+-------------+-----------+-------------+
| table_schema | column_name | data_type | is_nullable |
|    varchar   |   varchar   |  varchar  |   varchar   |
+--------------+-------------+-----------+-------------+
| public       | id          | Int64     | YES         |
| public       | name        | Utf8      | YES         |
| public       | properties  | Utf8      | YES         |
| public       | color       | Utf8      | YES         |
+--------------+-------------+-----------+-------------+

Time: 0.004333084 seconds. 4 rows.
```

The README previously showed both of these as three-column tables. What this PR writes into `json_strings/README.md` is byte-for-byte the output above.

The timestamp spelling, on the same build:

```console
sql> select arrow_typeof(arrow_cast(now(), 'Timestamp(Microsecond, Some("Etc/UTC"))')) as tz,
            arrow_typeof(arrow_cast(1, 'Timestamp(Microsecond, None)')) as notz;
+--------------------------+---------------+
|            tz            |      notz     |
|          varchar         |    varchar    |
+--------------------------+---------------+
| Timestamp(µs, "Etc/UTC") | Timestamp(µs) |
+--------------------------+---------------+

Time: 0.06061975 seconds. 1 rows.
```

And a four-column `describe` over a parquet dataset with a `Timestamp(Microsecond, None)` column, confirming the two render together as the `kubernetes` transcript now shows:

```console
sql> describe taxi_trips;
+--------------+-----------------------+---------------+-------------+
| table_schema |      column_name      |   data_type   | is_nullable |
|    varchar   |        varchar        |    varchar    |   varchar   |
+--------------+-----------------------+---------------+-------------+
| public       | VendorID              | Int32         | YES         |
| public       | tpep_pickup_datetime  | Timestamp(µs) | YES         |
| public       | tpep_dropoff_datetime | Timestamp(µs) | YES         |
…
```

### The other four recipes

`oracle`, `redshift`, `kubernetes` and `spark` each need a live Oracle / Redshift / Kubernetes / Spark backend, none of which was available here, so those four transcripts were **not re-run**. Their edit is mechanical and follows from the source above rather than from a run: the column list and ordering are fixed by the REPL's rewrite regardless of connector, the type row is the REPL's own renderer, and the timestamp spelling is the shared Arrow formatter shown above. Each table here was regenerated with a renderer checked byte-for-byte against the live v2.3.1 output quoted in this PR. The per-recipe column names and types were carried over unchanged, so nothing connector-specific is being asserted.

## Flip-flop guard

```console
$ gh pr list -R spiceai/cookbook --state merged --limit 40 --search "describe"
(no merged PR touches these describe transcripts)
```

Open PR #660 also edits `oracle`, `redshift`, `kubernetes` and `spark`, but only their `show tables` transcripts — its hunks are at oracle:88, redshift:261, kubernetes:70/130, spark:129, none of which overlap the `describe` blocks changed here.


grokspice:
@copilot review

lukekim (approved):

